PR #27, opened 2017-06-06
Remove comment about colorisation information

glguy:
I believe that this comment was a hold-over from the original forked library that had built-in colorization.


quchen:
Looks like it. Thanks!

Issue #30, opened 2017-06-24
Add Pretty instance for Doc


sergv:
It seems that in general such instance cannot be added due to annotation mismatch between input and output. However, limited version can be introduced and still would be useful:

```haskell
instance Pretty (Doc Void)
```

quchen:
The (Doc ann) instance was recently removed, since it had to traverse the entire Doc (with unAnnotate), which is quite a negative performance surprise. I don't think it's a good instance, do you actually have a good use case or did you just feel like it was missing?

sergv:
Well, to be honest, after trying to use it I think this instance is less useful than I initially planned. I was trying to port some utilities from `wl-pprint-text` package over to `prettyprinter` package. The utilities are organized in layers where basic functions like `ppList :: Pretty a => Doc b -> Doc b -> [a] -> Doc b` are used by other high-lever functions like `ppMap`. The aim before was to call `pretty` inside utility functions like `ppList` to make them suitable for use by both user and other more high-level utilities. E.g. user would pretty-print his own lists, while more high-level utilities would feed lists of already processed `Doc a` elements to `ppList` function. Without `Pretty (Doc a)` instance this didn't work, so I created this issue.

However, I think this use case is flawed and my helpers should be fixed instead. The problem with having `Pretty (Doc Void)` instance is that calls to `pretty` need to be explicitly annotated to produce `Doc Void` elements. There's also a problem of traversing entire `Doc` with `absurd` function (however this can be fixed via `unsafeCorerce` since we're guaranteed to have no annotations).

To summarise, it turns out I don't have a good use case for such instance nor I feel like it is missing.

quchen:
I'm okay with unsafeCoerce as long as it's safe. I've tried to get a good Pretty Doc instance to work, but none satisfied me :-| As for your use case, wouldn't a newtype with the desired instance do the trick as well? Anyway, thanks for the ticket and the patch :-)

PR #31, opened 2017-06-24
Add 'Pretty (Doc Void)' instance

sergv:
It's pretty useful to have `Pretty` instance for `Doc` so that parts of a complex structure can be printed independently and/or with different styles. Due to presence of annotations, adding more general instance
is problematic. However, this limited instance seems to be a good trade-off.

Addresses #30.


sergv:
This instance is not as useful as it might seem and has suboptimal complexity, as explained in #30.

PR #35, opened 2017-07-21
Fix cabal.project file and document SimpleDocTree

mpickering:


quchen:
Woops, forgot merging this! Sorry about that. Thanks for the contribution!

PR #38, opened 2017-08-17
Invert impl(ghc < ver) into !impl(ghc >= ver)

phadej:
See `impl` note in
http://cabal.readthedocs.io/en/latest/developing-packages.html?highlight=impl#conditions
for explanation.

> As a result, !impl(ghc >= x.y.z) is not entirely equivalent to
> impl(ghc < x.y.z). The test !impl(ghc >= x.y.z) checks that:
>
> * The current compiler is not GHC, or
>
> * The version of GHC is earlier than version x.y.z.


quchen:
Hah, I had no idea! Thanks for the patch.

Issue #39, opened 2017-08-22
Lifting Pretty to * -> * & * -> * -> *


robrix:
Would you be open to a PR defining liftings of `Pretty` to * -> * and * -> * -> *, a la `Show1` &c. from `Data.Functor.Classes`? It makes (decidable) recursive instances for `Fix`, `Free`, `Cofree` &c. much easier to define:

```haskell
instance Pretty1 f => Pretty (Fix f) where
  pretty (Fix f) = liftPretty pretty prettyList f
```

If so, I’ll be happy to submit one 😊

robrix:
For reference, here’s the implementation I’m working with currently:

https://github.com/robrix/abstract-interpretation/blob/master/src/Data/Functor/Classes/Pretty.hs

(I’d document everything before submission, of course.)

quchen:
Hm, the code doesn’t really look very nice, but since we now have Show1 etc. in Base, I think this is something we could include. I’d like to keep it minimally invasive though – from a Haddock standpoint, there shouldn’t be too many new exports, the (single exposed main) module is long enough as it is.

Can you explain why `Pretty1Of` is necessary a bit? I’m familiar with `Show1`, but not with `Show1Of`.

robrix:
> Hm, the code doesn’t really look very nice, but since we now have Show1 etc. in Base, I think this is something we could include. I’d like to keep it minimally invasive though – from a Haddock standpoint, there shouldn’t be too many new exports, the (single exposed main) module is long enough as it is.

Okay, you bet 👍

Would you be interested in including a generically-derivable implementation of `Pretty1`? This really isn’t make-or-break for me; we can always include it in our project and replace `deriving Pretty1` with `instance Pretty1 F where liftPretty = genericLiftPretty`.

I’m guessing not given the above, but wanted to double-check since it wouldn’t necessarily have to increase the API’s surface area.

> Can you explain why `Pretty1Of` is necessary a bit? I’m familiar with `Show1`, but not with `Show1Of`.

Oh, my bad—I don’t think we should include it at all. It’s a convenience to give you a `Pretty` instance if you have a `Pretty1` instance, and helps to avoid orphan instances. I see little benefit to including it here, because presumably we’d just write the `Pretty` instance in the first place.

quchen:
I don’t think Pretty1 should be derivable (likewise for Pretty), since prettiness is not something that follows any laws. If push comes to shove and we really need a quick Pretty instance, there is `prettyViaShow`.

[...]

PR #40, opened 2017-08-23
Lifted Pretty classes

robrix:
Per #39, this PR defines `Pretty1` & `Pretty2` classes, lifting `Pretty` to `* -> *` & `* -> * -> *` respectively, in the same manner as `Show1` & `Show2`.

I’ve also defined a few instances of these new classes, as well as an instance of `Pretty` for `Either a b`, since it seemed incomplete to provide a `Pretty2` instance for `(,)` but not for `Either`, and thereafter seemed incomplete to provide only lifted instances for `Either`.

I’ve confirmed that the tests (including doctests) all pass, but haven’t added any new tests since the instances are quite minimal.


prettyprinter/src/Data/Text/Prettyprint/Doc/Internal.hs
```diff
@@ -352,7 +361,95 @@ instance Pretty Lazy.Text where pretty = pretty . Lazy.toStrict
...
+        -> ([b] -> Doc ann) -- ^ A function to print a list of the second parameter.
+        -> f a b
+        -> Doc ann
+
+-- | Print 'Left' and 'Right' contents with the supplied functions.
+--
+-- >>> liftPretty2 (parens . pretty) (list . map (parens . pretty)) (parens . pretty) (list . map (parens . pretty)) (Left True :: Either Bool Bool)
+-- (True)
+-- >>> liftPretty2 (parens . pretty) (list . map (parens . pretty)) (parens . pretty) (list . map (parens . pretty)) (Right True :: Either Bool Bool)
+-- (True)
```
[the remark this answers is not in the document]
quchen: I think these tests would be nicer if they weren’t on a single line, but with `let` helper definitions with good names.

prettyprinter/src/Data/Text/Prettyprint/Doc/Internal.hs
```diff
@@ -352,7 +361,95 @@ instance Pretty Lazy.Text where pretty = pretty . Lazy.toStrict
...
+-- This is most useful for:
+-- 1. defining 'Pretty' instances for recursive types,
+-- 2. defining 'Pretty' instances for type constructors without 'Functor'
+--    instances, and
+-- 3. efficiently pretty-printing type constructors with 'Functor' instances
+--    whose 'fmap' traverses the whole structure.
+--
+-- Laws:
+--
+--   1. output should be pretty. :-)
```
quchen: I think there should be an example for an instance definition here, for example the `[]` one. And the law joke doesn’t need to be repeated again ;-)

I think the key to the Pretty1 documentation should be showing how it is useful without going into too much detail. Like »here we define Pretty without using Pretty1, see how Pretty1 makes our lives much easier?«

prettyprinter/src/Data/Text/Prettyprint/Doc/Internal.hs
```diff
@@ -352,7 +361,95 @@ instance Pretty Lazy.Text where pretty = pretty . Lazy.toStrict
...
+class Pretty1 f where
+
+    -- | >>> liftPretty (parens . pretty) (list . map (parens . pretty)) (Just "hello")
+    -- (hello)
+    liftPretty
+        :: (a -> Doc ann)   -- ^ A function to print a single value.
+        -> ([a] -> Doc ann) -- ^ A function to print a list. Used for [].
```
quchen: Is this really necessary? It clutters the definitions quite a bit, but it’s usually just `list . map pretty`.

robrix: It’s necessary for the `[]` definition to work as expected for `Char`, unfortunately.

quchen: Arrr, I see.

Actually, this brings up a new law: `Pretty1 f` and `Pretty (f a)` should result in identical behavior! :-)

robrix:
> I think the key to the Pretty1 documentation should be showing how it is useful without going into too much detail. Like »here we define Pretty without using Pretty1, see how Pretty1 makes our lives much easier?«

I quite like this idea! I think the `[]` instance isn’t gonna cut it, because it’s almost identical to the `Pretty` instance. Unfortunately, I’m having difficulty coming up with a better example. My own motivating examples tend to involve GADTs, signature functors, and/or functor composition, e.g.:

```haskell
{-# LANGUAGE GADTs #-}
data Expr a = Plus a a | Times a a | Const Int
newtype Fix f = Fix (f (Fix f))
data FreeF f a b = Free (f b) | Pure a

type Free f a = Fix (FreeF f a)

instance Pretty1 Expr where
  liftPretty p _  (Plus a b) = parens $ p a <+> pretty '+' <+> p b
  liftPretty p _  (Times a b) = p a <+> pretty '*' <+> p b
  liftPretty _ _ (Const i) = pretty i

instance Pretty1 f => Pretty2 (Free f) where
  liftPretty2 pA _ pB plB = go
    where go (Pure a) = pA a
          go (Free f) = liftPretty pB plB f

instance (Pretty1 f, Pretty a) => Pretty1 (Free f) where
  liftPretty = liftPretty2 pretty prettyList

instance Pretty1 f => Pretty (Fix f) where
  pretty (Fix f) = liftPretty pretty prettyList f
```

vs.

```haskell
{-# LANGUAGE UndecidableInstances #-}
instance (Pretty (f b), Pretty a) => Pretty (FreeF f a b) where
  pretty (Pure a) = pretty a
  pretty (Free f) = pretty f

instance Pretty (f (Fix f)) => Pretty (Fix f) where
  pretty (Fix f) = pretty f
```

This seems a little overboard for the docs, but I’m not sure what would be a better example :confused: What do you think?


> Actually, this brings up a new law: `Pretty1 f` and `Pretty (f a)` should result in identical behavior! :-)

I suppose that if we were willing to compromise on that by accepting strange behaviour with `String`:

```
>>> liftPretty pretty "hello"
[h, e, l, l, o]
```

then we’d be able to pare things down significantly. I am having trouble imagining a situation where you’d use `liftPretty` on a `String`, to be honest; but I’ll defer to your preference.

quchen:
Back from summer hiatus! Sorry for the late reply. I don’t think `String` is important enough to break a leg over it to be honest!

robrix:
> Back from summer hiatus! Sorry for the late reply.

No apologies necessary; hope you had a lovely break!

> I don’t think `String` is important enough to break a leg over it to be honest! A short remark about this misalignment would be a good idea anyway, since for example `Maybe` a has a special case for lists as well.

Done and done 👍

> Your documentation looks good, but it’s too verbose to show it in the default view. But good news, I recently found out that Haddock allows collapsible sections via `-- ==== <title>` ([example](https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Maybe.html#v:listToMaybe))!

Oh, very cool! Thanks for sharing that gem 😊

> Putting the long and somewhat complicated example in there would be fine: it allows interested parties to see them, while not being interrupting for the casual reader.

Excellent. It took me a bit to figure out how to make doctest accept multi-line input for the instances, but I’ve now added that and it’s looking good 👍

I appreciate you taking the time to workshop this PR with me. IMO the code has benefitted greatly from it! Let me know what you think of the latest revisions.

PR #54, opened 2018-03-21
Fix typos in documentation of renderStrict

cocreature:
Looks like a copy-paste gone wrong :)


quchen:
Woopsie, thanks for the patch!

PR #60, opened 2018-05-11
Include changelog of prettyprinter in extra-source-files

cocreature:
This will include it in the sdist tarball which means it shows up on Hackage. The other packages in this repository either already do this (in the case of `prettyprinter-ansi-terminal`) or don’t have changelog files (all other packages).

### DIFF
--- a/prettyprinter/prettyprinter.cabal
+++ b/prettyprinter/prettyprinter.cabal
@@ -7,6 +7,7 @@ description:         See README.md
 license:             BSD2
 license-file:        LICENSE.md
 extra-source-files:  README.md
+                   , CHANGELOG.md
                    , misc/version-compatibility-macros.h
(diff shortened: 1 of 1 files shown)


prettyprinter/prettyprinter.cabal
```diff
@@ -7,6 +7,7 @@ description:         See README.md
 license:             BSD2
 license-file:        LICENSE.md
 extra-source-files:  README.md
+                     CHANGELOG.md
```
quchen: Missing a comma here?

quchen:
Good catch! Is the missing comma (see comment) relevant?

cocreature:
The missing comma doesn’t seem to matter (I tested `cabal sdist` and it included all three files) but to keep things consistent I added the comma.

quchen:
Excellent, thanks for the patch!

PR #72, opened 2019-04-20
Fix .cabal file

vmchale:
Right now `prettyprinter` can't be cross-compiled even as a library, because the `QuasiQuotes` extension is marked in `other-extensions`. This moves it to the appropriate stanza, so that the library can be used.


quchen:
Thanks for the patch! Rebased and merged.

PR #74, opened 2019-05-06
Widen support to GHC-7.4...GHC-8.8

phadej:
- GHC-8.8 is straight-forward `MONAD_NO_FAIL` fix
- GHC-7.4 + GHC-7.6 is trickier
  - **motivation** We'd like to use `prettyprinter` in `trifecta` and `criterion`, and it would be nice to keep support for older GHCs (especially for `criterion`). In this case the price is not that big.
  - no `LambdaCase`
  - and `AutoDeriveTypeable` -> `DataDeriveTypeable`
  - there's small problem, `pgp-wordlist` is `base >=4.6` (not `base >= 4.5`), but with `--allow-older=pgp-wordlist:base` it compiles fine.

There are few follow-ups which I'd like to do:
- extend `pgp-wordlist` support to GHC-7.4 too (should be simple)
- use [`haskell-ci`](https://github.com/haskell-ci/haskell-ci) for CI, so we can test old GHCs (and new!)
- split benchmarks into own package, because if/when `criterion` would depend on `prettyprint` there would be nasty package loop. Similar thing is done in e.g. https://github.com/haskell/containers/pull/626

cc @RyanGlScott

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -26,6 +26,9 @@
 - ignore: {name: "Use first"}
 - ignore: {name: "Use second"}
 
+# Use LambdaCase -- we cannot, GHC-7.6+ feature
+- ignore: {name: "Use lambda-case"}
+
 # Not all 3-liners should be deduplicated
 - ignore: {name: Reduce duplication, within: Data.Text.Prettyprint.Doc.Render.Terminal.Internal}
 
--- a/aux/version-compatibility-macros.h
+++ b/aux/version-compatibility-macros.h
@@ -14,7 +14,8 @@
 #define MONOID_IN_PRELUDE               MIN_VERSION_base(4,8,0)
 #define NATURAL_IN_BASE                 MIN_VERSION_base(4,8,0)
 
-#define MONAD_FAIL                      MIN_VERSION_base(4,9,0)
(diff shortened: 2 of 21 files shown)


phadej:
Looks like travis is disabled for this and https://github.com/quchen/pgp-wordlist/pull/3

RyanGlScott:
I certainly appreciate the work that goes into PRs like this. Do note that if your eventual goal is to get `criterion` to use `prettyprinter`, the real challenge is going to be `optparse-applicative` (on top of which `criterion` builds its pretty-printed CLI `--help` output), which supports all the way back to GHC 7.0. (See also pcapriotti/optparse-applicative#273).

phadej:
In fact, it’s simple-ish to extend support back to 7.0 too; one would need to make defaults/Generics optional. Or use generic-deriving.

I’m happy to walk that path.

 Having wide GHC support is a ”there are no type system magic” indicator. Which I value for basic libs, like pretty printing. OTOH it cames with some CPP cost; but not terrible.

With haskell-ci it will be easy to setup CI even in that case.

phadej:
ping @quchen

quchen:
PGP-Wordlist is only a test dependency, and not an important one at that. I think it can easily be removed.

PR #75, opened 2019-05-17
Support GHC-7

phadej:
This is untested.

I'd recommend to try to compile `optparse-applicative` `prettyprinter` patch with this version with GHC-7.0.4, to see how it works out.

Major bumps in `-compat` packages is due `Data.Monoid.<>` changing to `Data.Semigroup.<>`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -6,5 +6,5 @@ packages: prettyprinter
         , prettyprinter-compat-ansi-wl-pprint
         , prettyprinter-convert-ansi-wl-pprint
         , prettyprinter-compat-annotated-wl-pprint
-tests: true
-benchmarks: true
+-- tests: true
+-- benchmarks: true
--- a/prettyprinter-ansi-terminal/src/Data/Text/Prettyprint/Doc/Render/Terminal/Internal.hs
+++ b/prettyprinter-ansi-terminal/src/Data/Text/Prettyprint/Doc/Render/Terminal/Internal.hs
@@ -123,15 +123,15 @@ renderLazy sdoc = runST (do
     let push x = modifySTRef' styleStackRef (x :)
         unsafePeek = readSTRef styleStackRef >>= \tok -> case tok of
             []  -> panicPeekedEmpty
-            x:_ -> pure x
+            x:_ -> return x
         unsafePop = readSTRef styleStackRef >>= \tok -> case tok of
             []   -> panicPeekedEmpty
-            x:xs -> writeSTRef styleStackRef xs >> pure x
+            x:xs -> writeSTRef styleStackRef xs >> return x
         writeOutput x = modifySTRef outputRef (<> x)
 
     let go = \sds -> case sds of
             SFail -> panicUncaughtFail
-            SEmpty -> pure ()
+            SEmpty -> return ()
             SChar c rest -> do
                 writeOutput (TLB.singleton c)
                 go rest
--- a/prettyprinter-compat-wl-pprint/prettyprinter-compat-wl-pprint.cabal
+++ b/prettyprinter-compat-wl-pprint/prettyprinter-compat-wl-pprint.cabal
@@ -1,5 +1,5 @@
 name:                prettyprinter-compat-wl-pprint
-version:             1.0.0.1
+version:             1.1
 cabal-version:       >= 1.10
 category:            User Interfaces, Text
 synopsis:            Drop-in compatibility package to migrate from »wl-pprint« to »prettyprinter«.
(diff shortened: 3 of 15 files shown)


prettyprinter-ansi-terminal/src/Data/Text/Prettyprint/Doc/Render/Terminal/Internal.hs
```diff
@@ -123,15 +123,15 @@ renderLazy sdoc = runST (do
     let push x = modifySTRef' styleStackRef (x :)
         unsafePeek = readSTRef styleStackRef >>= \tok -> case tok of
             []  -> panicPeekedEmpty
-            x:_ -> pure x
+            x:_ -> return x
```
quchen: I’d rather have a conditional import of Applicative instead of using `return`, in order to keep compatibility hacks in the imports rather than in the code.

prettyprinter-compat-wl-pprint/prettyprinter-compat-wl-pprint.cabal
```diff
@@ -1,5 +1,5 @@
 name:                prettyprinter-compat-wl-pprint
-version:             1.0.0.1
+version:             1.1
```
quchen: Why not 1.0.1, or even 1.0.0.2? Widening support should not introduce any breakages.

prettyprinter/src/Data/Text/Prettyprint/Doc/Render/Util/StackMachine.hs
```diff
@@ -72,11 +72,11 @@ renderSimplyDecorated text push pop = go []
     go _           SFail               = panicUncaughtFail
     go []          SEmpty              = mempty
     go (_:_)       SEmpty              = panicInputNotFullyConsumed
-    go stack       (SChar c rest)      = text (T.singleton c) <> go stack rest
-    go stack       (SText _l t rest)   = text t <> go stack rest
-    go stack       (SLine i rest)      = text (T.singleton '\n') <> text (T.replicate i " ") <> go stack rest
-    go stack       (SAnnPush ann rest) = push ann <> go (ann : stack) rest
-    go (ann:stack) (SAnnPop rest)      = pop ann <> go stack rest
+    go stack       (SChar c rest)      = text (T.singleton c) `mappend` go stack rest
```
quchen: Was `<>` not part of `Data.Monoid` back then? If not, I would prefer a conditional local definition,

```haskell
#if !(DIAMOND_OPERATOR_SUPPORTED)
(<>) = mappend
#endif
```

cabal.project
```diff
@@ -6,5 +6,5 @@ packages: prettyprinter
         , prettyprinter-compat-ansi-wl-pprint
         , prettyprinter-convert-ansi-wl-pprint
         , prettyprinter-compat-annotated-wl-pprint
-tests: true
-benchmarks: true
+-- tests: true
+-- benchmarks: true
```
quchen: I’m unfamiliar with the cabal.project file, so I’ll just trust you on this one :-)

prettyprinter-ansi-terminal/src/Data/Text/Prettyprint/Doc/Render/Terminal/Internal.hs
phadej: http://hackage.haskell.org/package/base-4.3.0.0/docs/Control-Monad-ST.html not `Applicative` in GHC-7.0 base (`base-orphans` probably fixes that though)

prettyprinter-compat-wl-pprint/prettyprinter-compat-wl-pprint.cabal
phadej: There is breaking change. The compats exported `Data.Monoid ((<>))` not, `Data.Semigroup ((<>))`.

prettyprinter/src/Data/Text/Prettyprint/Doc/Render/Util/StackMachine.hs
phadej: that `<>` would clash with semigroup's `<>`.

cabal.project
phadej: recent `tasty` or `criterion` don't have install-plans on GHC-7.0 and GHC-7.2; unfortunately `cabal.project` doesn't support conditionals; so we cannot write

```haskell
if impl(ghc >= 7.4)
  tests: true
  benchmarks: true
```

to make tests and benchmarks build by default with recent GHC.

Not having benchmarks on older GHCs is completely fine. Tests are trickier, but I'm optimistic that we can extrapolate from other GHC results that "if it compiles, it works". 

One could miss some instance or symbol, and we do miss `Generic` stuff for GHC-7.0, so I'd recommend to try if `optparse-applicative` patch works on GHC-7.0; so those instances aren't essential for common use

phadej:
I guess it's not worth it

PR #82, opened 2019-09-10
Correct lower bounds

phadej:
Also add upper bounds to `fail` and `void`.

I also made revisions to existing `prettyprinter` releases


phadej:
ping @quchen

quchen:
Merged. Better late than never! :-S

PR #202, opened 2021-07-22
Make text an optional dependency

Bodigrim:
I should have raised an issue first, but it is easier to demonstate what I'm proposing with a PR. I'm fine to have it closed/rejected, if we find it unpalatable.

I very much understand that it's a core value of `prettyprinter` to use `Text` and not `String`. However, there are cases when you cannot afford it: namely, when you are a `text` (or `bytestring`) developer ;) or work with a very bleeding-edge GHC. If in future `ansi-wl-pprint` (already deprecated) is to extinct and be replaced by `prettyprinter` in dependencies of `optparse-applicative`, `tasty` and `test-framework`, developers of low-level packages would have to deal with pretty nasty build dependencies. E. g., to run `bytestring` tests you'd have to rebuild `text`, then `prettyprinter`, then `optparse-applicative`, then `tasty` and all its providers. This is extremely disheartening, as I can tell from earlier experience. 

I was interested to check how difficult is to make `prettyprinter` fallback to `String`, if `text` is not available. It seems that just couple of lines and a Cabal flag would suffice. How do you feel about this?

### DIFF
--- a/.github/workflows/no-text.yml
+++ b/.github/workflows/no-text.yml
@@ -32,5 +32,5 @@ jobs:
       with:
         path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
         key: ${{ runner.os }}-${{ matrix.ghc }}
-    - name: Build
-      run: cabal build -f-text prettyprinter:prettyprinter
+    - name: Test
+      run: cabal test -f-text prettyprinter:test:testsuite
--- a/.github/workflows/no-text.yml
+++ b/.github/workflows/no-text.yml
@@ -32,5 +32,5 @@ jobs:
       with:
         path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
         key: ${{ runner.os }}-${{ matrix.ghc }}
-    - name: Test
-      run: cabal test -f-text prettyprinter:test:testsuite
+    - name: Build
+      run: cabal build -f-text prettyprinter:prettyprinter
--- a/.github/workflows/no-text.yml
+++ b/.github/workflows/no-text.yml
@@ -1,0 +1,36 @@
+name: no-text
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ['ubuntu-latest']
+        ghc: ['latest', '7.10']
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v2
+    - uses: haskell/actions/setup@v1
+      id: setup-haskell-cabal
+      with:
+        ghc-version: ${{ matrix.ghc }}
+    - name: Update cabal package database
+      run: cabal update
+    - uses: actions/cache@v2
+      name: Cache cabal stuff
+      with:
+        path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}
+    - name: Build
+      run: cabal build -f-text prettyprinter:prettyprinter
--- a/prettyprinter/prettyprinter.cabal
+++ b/prettyprinter/prettyprinter.cabal
@@ -1,6 +1,6 @@
 name:                prettyprinter
 version:             1.7.0
-cabal-version:       >= 1.10
+cabal-version:       2.0
 category:            User Interfaces, Text
 synopsis:            A modern, easy to use, well-documented, extensible pretty-printer.
 description:         A modern, easy to use, well-documented, extensible pretty-printer. For more see README.md
@@ -73,13 +73,7 @@ library
     if flag(text)
         build-depends: text >= 1.2
     else
-        -- A fake text package, emulating the same API, but backed by String
-        hs-source-dirs: src-text
-        other-modules:
-              Data.Text
-            , Data.Text.IO
-            , Data.Text.Lazy
-            , Data.Text.Lazy.Builder
+        build-depends: fake-text
 
     if !impl(ghc >= 7.6)
         build-depends: ghc-prim
@@ -92,7 +86,19 @@ library
     if !impl(ghc >= 7.10)
         build-depends: void >=0.4 && <0.8
 
-
+-- A fake text package, emulating the same API, but backed by String
+library fake-text
+    build-depends: base
+    ghc-options: -Wall -O2
+    default-language: Haskell2010
+    hs-source-dirs: src-text
+    exposed-modules:
+          Data.Text
+        , Data.Text.IO
+        , Data.Text.Lazy
+        , Data.Text.Lazy.Builder
+    if flag(text)
+        buildable: False
 
 Flag buildReadme
   Description: Build the readme generator
@@ -158,16 +164,16 @@ test-suite testsuite
         , tasty            >= 0.10
         , tasty-hunit      >= 0.9
         , tasty-quickcheck >= 0.8
-        , text
     ghc-options: -threaded -rtsopts -with-rtsopts=-N -Wall
     default-language: Haskell2010
 
     if !impl(ghc >= 8.0)
         build-depends: semigroups >= 0.6
 
-    if !flag(text)
-        buildable: False
-
+    if flag(text)
+        build-depends: text
+    else
+        build-depends: fake-text
 
 benchmark fusion
     type: exitcode-stdio-1.0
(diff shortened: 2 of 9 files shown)


sjakobi (commented):
Nice hack! :smiling_imp: I'm curious what @quchen will say! ;)

I'd like to see:

* [ ] Brief explanatory module headers for the contents of `src-text`.
* [ ] CI with `-text`. I think this can be achieved with a `haskell-ci` [constraint set](https://github.com/haskell-CI/haskell-ci/blob/877fdee16e62bfb235cadea8c0b08f32a429b45c/cabal.haskell-ci#L97-L115).

sjakobi:
Also, I hope that there are no expectations that `prettyprinter` maintainers will work on performance tuning the `-text` implementation.

Bodigrim:
I'm happy to supply CI config and additional comments and whatever :) if @quchen approves this approach. 

There is also an option to separate `src-text` into an internal library, so that one can run tests and benchmarks in `-text` configuration. I've implemented it first, but reverted back, because it is a tiny bit more intrusive.

prettyprinter/prettyprinter.cabal
```diff
@@ -88,6 +97,11 @@ Flag buildReadme
   Description: Build the readme generator
   Default:     False
 
+Flag text
+  Description: Use Text (recommended default), otherwise do not depend
+               on text package and fallback to String
```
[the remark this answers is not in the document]
Bodigrim: I think to avoid confusion it'd be better to disable these renderers when invoked with `text: False`.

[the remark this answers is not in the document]
sjakobi: Good point.

quchen:
Wow, this is some hack. If it helps adoption, I’m in favor. I think you should add documentation for the very much non-intuitive new modules, a simple paste of the ticket’s description as Haddock should suffice. It’s also important that this does not show up the official Haddock documentation or in standard builds, but that shouldn’t be the case by default anyway, right?

prettyprinter/src-text/Data/Text.hs
```diff
@@ -0,0 +1,38 @@
+{-# OPTIONS_GHC -Wno-missing-signatures #-}
+
+module Data.Text where
```
quchen, line 3: Add some documentation of why this file exists

Bodigrim:
@quchen @sjakobi I've added more comments + CI build.

.github/workflows/no-text.yml
```diff
@@ -0,0 +1,32 @@
...
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ['ubuntu-latest']
+        ghc: ['latest']
```
sjakobi, line 16: We might want better coverage of GHC versions, but we can add that if it turns out that the compatibility story is not as simple as we hope.

.github/workflows/no-text.yml
```diff
@@ -0,0 +1,32 @@
...
+        ghc-version: ${{ matrix.ghc }}
+    - name: Update cabal package database
+      run: cabal update
+    - uses: actions/cache@v2
+      name: Cache cabal stuff
+      with:
+        path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}
+    - name: Build
+      run: cabal build -f-text prettyprinter:prettyprinter
```
sjakobi, line 32: Running the tests would be nice, but that's also something that we can try once we feel a stronger need.

sjakobi (approved):

.github/workflows/no-text.yml
Bodigrim: I've added 7.10 to the matrix (this is the earliest available through `ghcup`) and checked 7.4 manually. This actually unveiled some issues (e. g., old `base` does not have `Data.List.uncons`), now fixed.

.github/workflows/no-text.yml
Bodigrim: Seems I found a way to wire in tests without too much churn, pushed.

prettyprinter/prettyprinter.cabal
```diff
@@ -69,7 +67,22 @@ library
 
     build-depends:
           base >= 4.5 && < 5
-        , text >= 1.2
+
+    -- While it's a core value of prettyprinter to use Text, there are rare circumstances
+    -- (mostly when prettyprinter arises as a dependency of test suite of packages
+    -- like bytestring or text) when this is inconvenient. That's why we provide
+    -- a Cabal flag to use String instead and do not depend on text.
```
sjakobi, line 74: I think these docs should be placed next to the `Flag text` declaration.

prettyprinter/prettyprinter.cabal
```diff
@@ -82,12 +95,29 @@ library
     if !impl(ghc >= 7.10)
         build-depends: void >=0.4 && <0.8
 
-
+-- A fake text package, emulating the same API, but backed by String
+library fake-text
```
sjakobi, line 99: I'm surprised that this internal library doesn't produce a warning in combination with `cabal-version: >= 1.10`. Has Cabal supported these for so long?! And `stack` can handle them too?!

(I'm just asking – I've never knowingly used an internal library before.)

prettyprinter/test/Testsuite/Main.hs
```diff
@@ -138,7 +142,7 @@ content = frequency
     [ (1, pure emptyDoc)
     , (10, do word <- choose (minBound, maxBound :: Word8)
               let pgp8Word = toText (BSL.singleton word)
-              pure (pretty pgp8Word) )
+              pure (pretty (read (show pgp8Word) :: T.Text)) )
```
sjakobi, line 145: I don't understand why the `read . show` bit is necessary. Could it be simplified?!

.github/workflows/no-text.yml
```diff
@@ -0,0 +1,32 @@
+name: no-text
+on:
+  - push
+  - pull_request
```
sjakobi, line 4: Copied from the `haskell-ci` setup:

```suggestion
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

sjakobi (commented):
Looks pretty good! A few more wibbles…

prettyprinter/prettyprinter.cabal
Bodigrim: Ah, `cabal build` was silent, but `cabal check` actually mentions that this requires `cabal-version: 2.0`. Do you think it is a problem? I can revert running tests in `-f-text` mode.

AFAICT `stack` can handle it well, I used this feature in other packages for some time.

prettyprinter/test/Testsuite/Main.hs
Bodigrim: `toText` above comes from `pgp-wordlist` package, so `pgp8Word` is a genuine `Text`, irrespective of `prettyprinter` flags. If `prettyprinter` is built with `text`, then `read . show` is just `id` here, but if `text` is disabled, `read . show` converts a genuine `Text` to a fake one. This is the simplest workaround, avoiding CPP.

prettyprinter/prettyprinter.cabal
sjakobi: I think we'll probably be fine with requiring `cabal-version: 2.0`. If it does turn out to cause trouble, we can still restore compatibility with `Cabal-1.10`.

prettyprinter/test/Testsuite/Main.hs
sjakobi: Oof, somehow this makes my head spin. I think I might prefer the CPP. Or define `textToText = read . show` and add an explanatory comment…

sjakobi:
The previous CI run had timed out for some reason. I've restarted it now.

Bodigrim:
@sjakobi There appears to be a subtle issue with Stack support of internal libraries: while `stack build && stack repl` works, `rm -rf .stack-work && stack repl` does not. Unfortunately, from what I can gather, this seems to be the same kind of issues that hinder Backpack support in Stack and unlikely to be resolved any time soon. So I reverted a commit, covering test support (let it remain in git history, if someone in future will be interested to give it another go).

Otherwise this is ready for a final review, I believe.

sjakobi (approved):
@Bodigrim Ah, that's unfortunate.

@quchen Let's merge this PR by rebasing or with a merge commit to preserve the commit history.

quchen (approved):

sjakobi:
Thank you, @Bodigrim! :) I'll try to start working on a release this week.

PR #229, opened 2022-07-04
Reword misleading documentation for layoutPretty

benjamin-hodgson:
The docs implied that `layoutPretty` looks ahead by a single text element, which is not accurate - it looks ahead to the end of the line. I also reworded the `layoutSmart` docs, which implied that it tries to fit the whole document into one line.


sjakobi (approved):
Thanks!

Issue #236, opened 2023-01-19
Use `Foldable` instead of list?


andreasabel:
I see that some functions, like [`concatWith`](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:concatWith) work for any `Foldable` instead of just for list.

Is there a reason why the other list functions are not generalized to `Foldable`?  E.g. 
> hsep :: [[Doc](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#t:Doc) ann] -> [Doc](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#t:Doc) ann

I am trying to be progressive, using `Data.List.NonEmpty` and other list variants in my code base, but then I always have to sprinkle an extra `toList` in.

Unless there are reasons (like breaking type inference), I think all such list functions should be generalized to `Foldable`.  (Even if type inference is broken for legacy users, such a generalization could maybe be pushed in a major-major version bump like 2.0...)

sjakobi:
@a-02 Thoughts on this?

a-02:
As much as I appreciate people using more robust polymorphic containers, their use comes with the cost of using a big `base` typeclass like `Foldable` that we can't account for all the instances of. @sjakobi brought up a great point about `Foldable`s instances for Either, Maybe, and `(,) a`, in that `hsep` wasn't meant to be used in those contexts. 

On the other hand, the `O(n)` of `vector`s `toList` absolutely sucks and I would hate if our users had to call it even once. My dream implementation would be seperate `PrettyPrinter.Internal.Vector` and `PrettyPrinter.Internal.NonEmpty` modules, but that's probably overkill. Switching everything downstream of `concatWith` to use `Foldable` is fine honestly. Much more upside than downside.

sjakobi:
Alright, let's do it. #240 does at least some of the necessary conversions – not sure if anything else also needs generalization to `Foldable`.

PR #240, opened 2023-04-29
let hsep, vsep, fillSep, sep, hcat, vcat, fillCat and cat for any Foldable

KommuSoft:


sjakobi:
These changes should be discussed in #236 first.

sjakobi (approved):
Thanks!

I guess we could consider generalizing `encloseSep`, `list` and `tupled` too, but that doesn't need to happen in this PR.

PR #244, opened 2023-05-26
Change layout function called for renderPretty in ansi-wl-pprint compat

HuwCampbell:
Layout smart is more expensive, and can potentially hang in situations
where layout pretty works fine.


HuwCampbell:
Ping

Please merge this and release a patch branch. There's a reasonable chance that people's CLIs will hang with 0.17.1 of optparse applicative because of this.

HuwCampbell:
Hello. Just a bump on this.

sjakobi (approved):
Yes, I think this is a bug introduced in https://github.com/quchen/prettyprinter/commit/897e9a71a9ade41ff573074b7174221972f50941.

PR #256, opened 2024-11-04
Add PrettyAnn type class

### LINKED ISSUE #222: Proposal: Add PrettyAnn typeclass
As discussed in #102, there is currently no way to use the `Pretty` typeclass in combination with annotations. In that issue, several alternatives were discussed, which were either backwards incompatible or otherwise potentially computationally expensive.
But the first issue also mentioned the "obvious" version of the `Pretty` typeclass. Namely:

```haskell
class Pretty a ann where
    pretty :: a -> Doc ann
```

But changing the definition of the typeclass would be obviously backwards incompatible.
So I propose the following change:

## Add a PrettyAnn typeclass

Add the following typeclass to the library

```haskell
class PrettyAnn a ann where
    prettyAnn :: a -> Doc ann
```

together with the obvious instances for primitives, strings, text etc. Just like for `Pretty` right now.

I see the following benefits:

- Purely additive change, no problems with backwards incompatibility.
- "Simple" solution. Neither fancy types nor other typeclasses are involved. Only MultiParamTypeclasses.
- Performance characteristic is obvious: No implicit insertions of `unAnnotate` or similar functions which traverse the Doc.
- This is already a solution users implement downstream in their applications. Why not provide a canonical way how to do it?

And the following downsides:

- Increases API surface. Now there are two typeclasses for prettyprinting. The documentation could just explicitly state that the user should use `Pretty` for prettyprinting without annotations, and `PrettyAnn` for prettyprinting with annotations.
- Does not lead to the "ideal" solution, since the `Pretty` class is left as a historical wart, instead of being a special case of `PrettyAnn a Void`

mmhat:
This PR adds a `PrettyAnn` type class that supports embedding annotations in the `Doc`.

Fixes #222


prettyprinter/src/Prettyprinter/Internal.hs
```diff
@@ -347,6 +350,54 @@ instance Pretty Char where
     prettyList = vsep . map unsafeTextWithoutNewlines . T.splitOn "\n"
 #endif
 
+-- | This class is similar to 'Pretty', but allows you to embed annotations in
+-- the 'Doc'.
+--
+-- @since 1.7.1
+class PrettyAnn a ann where
```
andreasabel, line 357: To me it seems that the parameter should be ordered the other way round:
````haskell
class PrettyAnn ann a
````
In typical uses, I imagine the user chooses a fixed `ann` and then implements `PrettyAnn` instances for various `a`s.
It feels more natural to have the fixed parameter first and then the variable one.

For example see all the instances you define below.  
You would get a natural alignment for all the
```haskell
instance PrettyAnn ann ...
```

mmhat: @andreasabel Thanks for the feedback! You're absolutely right; That is better!

prettyprinter/src/Prettyprinter/Internal.hs
```diff
@@ -347,6 +350,54 @@ instance Pretty Char where
     prettyList = vsep . map unsafeTextWithoutNewlines . T.splitOn "\n"
 #endif
 
+-- | This class is similar to 'Pretty', but allows you to embed annotations in
+-- the 'Doc'.
+--
+-- @since 1.7.1
+class PrettyAnn ann a where
```
sjakobi: What's the reason for the swapped order of type parameters compared to the proposal in #222?

mmhat: IIRC I wanted the `a` to be the last type parameter since it is the last type parameter for `Pretty` too; Hence given a fixed annotation type the signatures align:
```haskell
Pretty :: Type -> Constraint
PrettyAnn MyAnnotation :: Type -> Constraint
```
Then it is also easier to define aliases like
```haskell
type MyPretty = PrettyAnn MyAnnotation
```

Not a very strong argument, I know; Should I change that too?

sjakobi: No, it's fine. I just saw that @andreasabel argued in favor of this order of parameters too: https://github.com/quchen/prettyprinter/pull/256#discussion_r1878633783

ony (approved):

PR #260, opened 2025-10-03
Add support for "faint" intensity

pbrisbin:
The "faint" console intensity (`\ESC[2m`) makes for a nice dim color and
is useful for rendering comments or other, less important text.

For example, the body of the PR being created by my script here:

The `ansi-terminal-types` documentation says it is "not widely
supported"[^1], which may be why it's not supported here yet, but I've
found other sources that say the same thing of italics[^2]. Besides,
folks can simply choose not to use.

[^1]: https://hackage-content.haskell.org/package/ansi-terminal-types-1.1.3/docs/System-Console-ANSI-Types.html#t:ConsoleIntensity
[^2]: https://stackoverflow.com/a/33206814


pbrisbin:
Closing in favor of #224.

PR #263, opened 2025-11-29
Try to fix CI

sjakobi:


sjakobi:
I've merged this to get CI working again.

There are some things that should probably be fixed before the next release though.

sjakobi:
I don't mind that I've had to remove CI with GHC < 8 though.

PR #285, opened 2026-04-14
Remove compatibility code for GHC < 8

sjakobi:
Follow-up to 92cc704 (drop GHC 7 support). Remove code that was only
    needed for GHC versions older than 8:
    
    - Remove CPP macros always true with base >= 4.9:
      APPLICATIVE_MONAD, FUNCTOR_IDENTITY_IN_BASE, MONOID_IN_PRELUDE,
      NATURAL_IN_BASE, SEMIGROUP_IN_BASE
    
    - Remove conditional dependencies on semigroups, fail, void, ghc-prim
    
    - Always enable -Wcompat (was conditional on GHC >= 8.0)
    
    - Remove doctest buildability conditions for GHC < 7.10
    
    - Remove unused Control.Applicative imports (in Prelude since base 4.8)
    
    Assisted-by: GLM-5 <noreply@fireworks.ai>

### DIFF
--- a/prettyprinter/bench/Fusion.hs
+++ b/prettyprinter/bench/Fusion.hs
@@ -18,10 +18,6 @@ import           Prettyprinter
 import           Prettyprinter.Render.Text
 import qualified Text.PrettyPrint.ANSI.Leijen as WL
 
-#if !(APPLICATIVE_MONAD)
-import Control.Applicative
-#endif
-
 
 
 main :: IO ()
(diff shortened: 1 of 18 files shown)


prettyprinter/bench/Fusion.hs
```diff
@@ -47,6 +43,8 @@ benchOptimize = env randomShortWords benchmark_
     randomShortWords' :: Int -> State StdGen [Text]
     randomShortWords' n = replicateM n randomShortWord
 
+    -- Note: 'split' is deprecated in random >= 1.2 in favor of 'splitGen'.
+    -- We could use 'splitGen' when support for GHC < 9.0 is dropped.
```
a-02, line 47: Any timeline on this?

a-02 (commented):
Looks a lot cleaner. Less stuff to worry about is always good.

prettyprinter/bench/Fusion.hs
sjakobi: Not really. I would like to keep support for GHC 8 at least for the 1.8 release.

When keeping the compatibility becomes very onerous I don't mind dropping it.

a-02: Fair, support for <9.0 isn't too hard (at the moment).

